PR #444, opened 2023-07-11
Remove unused variables // adam & adamw

Frightera:
`beta_1_power` and `beta_2_power` are defined in the next lines, so they can be removed from the beginning of the method.


fchollet (approved):
LGTM, thanks!

PR #449, opened 2023-07-11
Add names to the input tensor, small typo

Frightera:
If we don't specify the names in the input tensor, we get:

```python
ValueError: When providing `inputs` as a dict, all keys in the dict must match the names of
the corresponding tensors. Received key 'x1' mapping to value 
<KerasTensor shape=(None, 10), dtype=float32, name=keras_tensor_16> which has name
'keras_tensor_16'. Change the tensor name to 'x1' (via `Input(..., name='x1')`)
```

It works if we specify the names as suggested in the error.

### DIFF
--- a/keras_core/models/functional.py
+++ b/keras_core/models/functional.py
@@ -31,10 +31,10 @@ class Functional(Function, Model):
     Example:
 
     ```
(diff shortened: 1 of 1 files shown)


fchollet (approved):
LGTM, thanks!

Issue #455, opened 2023-07-12
ValueError: Unable to import backend : pytorch


ahmadmustafaanis:
I can not import backend PyTorch on Colab.
```python
!pip install -q keras-core

import torch
print(torch.__version__)  # 2.0.1+cu118

# set backend for Keras core

import os
os.environ["KERAS_BACKEND"] = "pytorch"  # gives same output on backend = torch

import importlib
import torch
import keras_core
from keras_core import backend as K

print(K.backend())
```

The output is
```
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
[<ipython-input-3-1420b7863060>](https://localhost:8080/#) in <cell line: 8>()
      6 import importlib
      7 import torch
----> 8 import keras_core
      9 from keras_core import backend as K
     10 

5 frames
[/usr/local/lib/python3.10/dist-packages/keras_core/__init__.py](https://localhost:8080/#) in <module>
      6 
      7 
----> 8 from keras_core import activations
      9 from keras_core import applications
     10 from keras_core import backend

[/usr/local/lib/python3.10/dist-packages/keras_core/activations/__init__.py](https://localhost:8080/#) in <module>
      6 
      7 
----> 8 from keras_core.src.activations import deserialize
      9 from keras_core.src.activations import get
     10 from keras_core.src.activations import serialize

[/usr/local/lib/python3.10/dist-packages/keras_core/src/__init__.py](https://localhost:8080/#) in <module>
----> 1 from keras_core.src import activations
      2 from keras_core.src import applications
      3 from keras_core.src import backend
      4 from keras_core.src import constraints
      5 from keras_core.src import datasets

[/usr/local/lib/python3.10/dist-packages/keras_core/src/activations/__init__.py](https://localhost:8080/#) in <module>
      1 import types
      2 
----> 3 from keras_core.src.activations.activations import elu
      4 from keras_core.src.activations.activations import exponential
      5 from keras_core.src.activations.activations import gelu

[/usr/local/lib/python3.10/dist-packages/keras_core/src/activations/activations.py](https://localhost:8080/#) in <module>
----> 1 from keras_core.src import backend
      2 from keras_core.src import ops
      3 from keras_core.src.api_export import keras_core_export
      4 
      5 

[/usr/local/lib/python3.10/dist-packages/keras_core/src/backend/__init__.py](https://localhost:8080/#) in <module>
     39     from keras_core.src.backend.torch import *  # noqa: F403
     40 else:
[...]

PR #457, opened 2023-07-12
Support "torch" and "pytorch" in Backend 

ahmadmustafaanis:
Since Torch is famous as PyTorch and Torch, there should be option to treat both as similar. ref: https://github.com/keras-team/keras-core/issues/455

### DIFF
--- a/keras_core/backend/__init__.py
+++ b/keras_core/backend/__init__.py
@@ -1,6 +1,6 @@
 from keras_core.backend.config import backend
 
-if backend() == "torch":
+if backend() in ["torch", "pytorch"]:
     # When using the torch backend,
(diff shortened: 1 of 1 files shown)


keras_core/backend/__init__.py
```diff
@@ -1,6 +1,6 @@
 from keras_core.backend.config import backend
 
-if backend() == "torch":
+if backend() in ["torch", "pytorch"]:
```
innat, line 3: How about only `pytorch`.

fchollet (commented):
Thanks for the PR. We're going to stick to a single identifier, and that should be `torch` which is the name of the package.

ahmadmustafaanis:
Thanks for the great work and reply @fchollet

PR #470, opened 2023-07-13
Export DTypePolicy as Policy for backward compat

abheesht17:
Resolves https://github.com/keras-team/keras-core/issues/468

### DIFF
--- a/keras_core/mixed_precision/__init__.py
+++ b/keras_core/mixed_precision/__init__.py
@@ -1,4 +1,5 @@
 from keras_core.mixed_precision.dtype_policy import DTypePolicy
+from keras_core.mixed_precision.dtype_policy import DTypePolicy as Policy
 from keras_core.mixed_precision.dtype_policy import dtype_policy
 from keras_core.mixed_precision.dtype_policy import set_dtype_policy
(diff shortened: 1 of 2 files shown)


abheesht17:
Does `@keras_core_export` not work as expected? Had to add an extra statement in `__init__.py`. There are other such cases in the repo. For example,

https://github.com/keras-team/keras-core/blob/d85be712db84b2c030a4e79a217c8066f08f08b1/keras_core/backend/config.py#L35-L37

```
>>> import keras_core as keras
Using TensorFlow backend
>>> keras.config.set_floatx
AttributeError: module 'keras_core' has no attribute 'config'
```

mattdangerw:
Two reasons you might be seeing `'keras_core' has no attribute 'config'`...

1) you did a `pip install .` or `pip install repo_url` instead of using `python pip_build.py --install`
2) you are running a script from the base of the keras-core repo, in which case the local directory `keras_core` will take precedence over the installed version.

abheesht17:
Ah, yes. You are right. The second reason is why it's happening

fchollet (approved):
LGTM, thanks

PR #501, opened 2023-07-16
Testing Timeseries anomaly detection Framework agnostic

anas-rz:
Tested this example for all frameworks. It can be placed under framework agnostic examples.
[Colab Notebook](https://colab.research.google.com/drive/1535mF70h17z3UH56EpEgbssEy9CQBkbE?usp=sharing)


fchollet (approved):
LGTM, thanks!

PR #502, opened 2023-07-16
Converted to Keras Core by: Yash Srivastava

yash-srivastava19:
I've converted the following [example](https://keras.io/examples/nlp/text_classification_with_transformer/) in the NLP section to Keras Core. It is backend agnostic(running succesfully on Jax, Pytorch and tf backend). 

There is, however a thing I need some help with - In the Transformer Embedding layer(`TokenAndPositionEmbedding`) , while adding the positional embeddings, I've done the following : 

``` positions = tf.range(start=0, limit = maxlen, delta = 1)```

As there was no function called `range` in `keras_core.ops`. Is it correct, or am I missing something?

### CODE REVIEW

File: examples/keras_io/nlp/text_classification_with_transformers.py [hunk 1]
```diff
@@ -0,0 +1,115 @@
+"""
+Title: Text Classification with Transformers
+Author: [Yash Srivastava](https://twitter.com/Yaaaaaashhh)
+Date created: 2023/07/16
+Last modified: 2023/07/16
```
Frightera, line 5: If this is not written from scratch, I guess the original date needs to be kept. Check: https://github.com/keras-team/keras-core/pull/491#discussion_r1264493629

examples/keras_io/nlp/text_classification_with_transformers.py
```diff
@@ -0,0 +1,115 @@
...
+Description: Implement a Transformer block as a Keras core layer and use it for text classification(Works on all backends).
+Accelerator: GPU
+"""
+"""
+## Setup
+"""
+
+import os
+os.environ['KERAS_BACKEND'] = 'jax'   # Could be anything out of 'jax', 'torch' or 'tf'
```
Frightera, line 14: TensorFlow backend is identified as `tensorflow`, not as `tf`

examples/keras_io/nlp/text_classification_with_transformers.py
```diff
@@ -0,0 +1,115 @@
...
+        super().__init__()
+        self.token_emb = keras.layers.Embedding(input_dim=vocab_size, output_dim=embed_dim)
+        self.pos_emb = keras.layers.Embedding(input_dim=maxlen, output_dim=embed_dim)
+
+    def build(self, input_shape):    # TODO: Check.
+        pass
+        
+    def call(self, x):
+        maxlen = keras.ops.shape(x)[-1]    # keras.ops has built in numpy and other nn related functions.
+        positions = tf.range(start=0, limit=maxlen, delta=1)   # TODO: Check if range is there or not.
```
Frightera, line 63: You can use `keras_core.ops.arange`.

examples/keras_io/nlp/text_classification_with_transformers.py
yash-srivastava19: Ok. Sure. I'll change this.

examples/keras_io/nlp/text_classification_with_transformers.py
yash-srivastava19: I'll update the comment as well.

examples/keras_io/nlp/text_classification_with_transformers.py
yash-srivastava19: Aah. This is something I missed. On it.

yash-srivastava19:
I've added the `keras_core.ops.arange()` (and changed the code accordingly), and also have corrected the things mentioned. I've removed the `build`, and added the `compute_output_shape` (which could be the bottleneck). Any pointers for the `compute_output_shape` function (if that is incorrect or needs to be done in a different way) would be really helpful.

examples/keras_io/nlp/text_classification_with_transformers.py
```diff
@@ -0,0 +1,115 @@
+"""
+Title: Text Classification with Transformers
+Author: [Yash Srivastava](https://twitter.com/Yaaaaaashhh)
```
Frightera, line 3: Please keep the original author, you can check the 4th point in Stage 1: https://github.com/keras-team/keras-core/issues/222

examples/keras_io/nlp/text_classification_with_transformers.py
```diff
@@ -0,0 +1,115 @@
...
+Date created: 2020/05/10              # Added the original date
```
Frightera, line 4: Remove the comment from here

examples/keras_io/nlp/text_classification_with_transformers.py
```diff
@@ -0,0 +1,115 @@
...
+Accelerator: GPU
+"""
+"""
+## Setup
+"""
+
+import os
+os.environ['KERAS_BACKEND'] = 'jax'   # Could be anything out of 'jax', 'torch' or 'tensorflow'   # Changed from 'tf' to tensorflow
+
+# import tensorflow as tf    # No need for importing tf, as keras_core handles it
```
Frightera, line 16: Remove the import, no need to reference TF here.

examples/keras_io/nlp/text_classification_with_transformers.py
```diff
@@ -0,0 +1,115 @@
...
+        self.att = keras.layers.MultiHeadAttention(num_heads=num_heads, key_dim=embed_dim)
+        self.ffn = keras.Sequential(
+            [keras.layers.Dense(ff_dim, activation="relu"), keras.layers.Dense(embed_dim),]
+        )
+        self.layernorm1 = keras.layers.LayerNormalization(epsilon=1e-6)
+        self.layernorm2 = keras.layers.LayerNormalization(epsilon=1e-6)
+        self.dropout1 = keras.layers.Dropout(rate)
+        self.dropout2 = keras.layers.Dropout(rate)
+
+    def compute_output_shape(self, input_shape):    # TODO: Check this. Could be a problem.
```
Frightera, line 35: Is that method necessary? I guess the original guide should stay the same, if possible.

examples/keras_io/nlp/text_classification_with_transformers.py
```diff
@@ -0,0 +1,115 @@
...
+        super().__init__()
+        self.token_emb = keras.layers.Embedding(input_dim=vocab_size, output_dim=embed_dim)
+        self.pos_emb = keras.layers.Embedding(input_dim=maxlen, output_dim=embed_dim)
+
+    def compute_output_shape(self, input_shape):   # TODO: Check this. Could be a problem
+      return (input_shape[-1], self.embed_dim)
+        
+    def call(self, x):
+        maxlen = keras.ops.shape(x)[-1]    # keras.ops has built in numpy and other nn related functions.
+        positions = keras.ops.arange(start=0, stop=maxlen, step=1)   # Changed from `tf.range` to `keras.ops.arange`, which removes the dependency of tf completely.
```
Frightera, line 63: I guess we can remove the comment from here.

Frightera (commented):
Thanks for the update, before any Keras team members reviews this, I left some more comments.

examples/keras_io/nlp/text_classification_with_transformers.py
yash-srivastava19: I tried having the same code of what was followed by the guide, but I encountered the error where it said that it wasn't able to understand the output shape.  A workaround for that was to explicitly implement what will be the output shape, which I did using the `compute_output_shape()` function.

examples/keras_io/nlp/text_classification_with_transformers.py
yash-srivastava19: Removed in the latest commit.

examples/keras_io/nlp/text_classification_with_transformers.py
yash-srivastava19: Removed in the latest commit.

examples/keras_io/nlp/text_classification_with_transformers.py
yash-srivastava19: I think I misread that. I'll add the original author's name.

examples/keras_io/nlp/text_classification_with_transformers.py
yash-srivastava19: Sure. Removed the comment in the latest commit.

examples/keras_io/nlp/text_classification_with_transformers.py
yash-srivastava19: The error that I encountered : 

``` Could not automatically infer the output shape / dtype of 'transformer_block' (of type TransformerBlock). Either the `TransformerBlock.call()` method is incorrect, or you need to implement the `TransformerBlock.compute_output_spec() / compute_output_shape()` method. ```

The `call` method is similar to the one used in the tutorial, so I implemented the `compute_output_shape`

**Comment by yash-srivastava19** [comment]
During call to `model_summary()`, I noticed my version has very few trainable parameters. Did I miss something? or do I need to pass a keyword argument like `trainable = True` for the layers? . Anything I missed ?(or do I need to implement the `build` function as well) ?

image


So less trainable parameters means definitely we are underfitting.

File: examples/keras_io/nlp/text_classification_with_transformers.py [hunk 9]
```diff
@@ -0,0 +1,113 @@
...
+"""
+
+class TokenAndPositionEmbedding(keras.layers.Layer):
+    def __init__(self, maxlen, vocab_size, embed_dim):
+        super().__init__()
+        self.token_emb = keras.layers.Embedding(input_dim=vocab_size, output_dim=embed_dim)
+        self.pos_emb = keras.layers.Embedding(input_dim=maxlen, output_dim=embed_dim)
+
+    def compute_output_shape(self, input_shape):   # TODO: Check this.
+      return (input_shape[-1], self.embed_dim)
```
fchollet, line 57: This is going to raise an error (the attribute doesn't exist).

examples/keras_io/nlp/text_classification_with_transformers.py
```diff
@@ -0,0 +1,113 @@
...
+Two seperate embedding layers, one for tokens, one for token index (positions).
+"""
+
+class TokenAndPositionEmbedding(keras.layers.Layer):
+    def __init__(self, maxlen, vocab_size, embed_dim):
+        super().__init__()
+        self.token_emb = keras.layers.Embedding(input_dim=vocab_size, output_dim=embed_dim)
+        self.pos_emb = keras.layers.Embedding(input_dim=maxlen, output_dim=embed_dim)
+
+    def compute_output_shape(self, input_shape):   # TODO: Check this.
```
fchollet, line 56: Implementing this method shouldn't be necessary.

examples/keras_io/nlp/text_classification_with_transformers.py
```diff
@@ -0,0 +1,113 @@
...
+        self.att = keras.layers.MultiHeadAttention(num_heads=num_heads, key_dim=embed_dim)
+        self.ffn = keras.Sequential(
+            [keras.layers.Dense(ff_dim, activation="relu"), keras.layers.Dense(embed_dim),]
+        )
+        self.layernorm1 = keras.layers.LayerNormalization(epsilon=1e-6)
+        self.layernorm2 = keras.layers.LayerNormalization(epsilon=1e-6)
+        self.dropout1 = keras.layers.Dropout(rate)
+        self.dropout2 = keras.layers.Dropout(rate)
+
+    def compute_output_shape(self, input_shape):    # TODO: Check this.
```
fchollet, line 33: Shouldn't be necessary

examples/keras_io/nlp/text_classification_with_transformers.py
```diff
@@ -0,0 +1,113 @@
...
+      return (input_shape[-1], self.embed_dim, self.embed_dim)
+
+    def call(self, inputs, training):
```
fchollet, line 36: Pretty sure this won't work, there's a shape mismatch

yash-srivastava19: Ok. I will check and all comments and let you know.

yash-srivastava19:
I've removed all the `compute_output_shape` methods, and the tutorial seems to be running fine. Turns out, previously I was using   a different version of keras_core, and in this commit, removed those methods. Check once on your end, and I can move to port other examples as well :) 

Thanks a lot being patient with me, as this was my first PR :)

yash-srivastava19:
> During call to `model_summary()`, I noticed my version has very few trainable parameters. Did I miss something? or do I need to pass a keyword argument like `trainable = True` for the layers? . Anything I missed ?(or do I need to implement the `build` function as well) ?
> 
> image
> 
> So less trainable parameters means definitely we are underfitting.

This was also resolved in the latest commit

fchollet:
The code doesn't seem to run for me? How did you test it?

yash-srivastava19:
In my colab environment, it worked without any errors, for all backends - without any errors. The version of keras_core was 0.1.0, for the commit(560ca2f70a215e17cb6f98d52cd13efb8368a412) : that was there 11 hours ago. Can you tell what is the error or I missed something again?

(I did install this using `pip install git+https://github.com/keras-team/keras-core.git` if that helps or not)

examples/keras_io/nlp/text_classification_with_transformers.py
```diff
@@ -0,0 +1,107 @@
...
+x = transformer_block(x)
+x = keras.layers.GlobalAveragePooling1D()(x)
+x = keras.layers.Dropout(0.1)(x)
+x = keras.layers.Dense(20, activation="relu")(x)
+x = keras.layers.Dropout(0.1)(x)
+outputs = keras.layers.Dense(2, activation="softmax")(x)
+
+model = keras.Model(inputs=inputs, outputs=outputs)
+
+print(model)
```
fchollet, line 94: This should be `model.summary()`

yash-srivastava19:
I just checked. The latest commit(cf87ee24ff306530b71c504812cd4e24aef521f8) is breaking at the transformer block.

yash-srivastava19:
> I just checked. The latest commit([cf87ee2](https://github.com/keras-team/keras-core/commit/cf87ee24ff306530b71c504812cd4e24aef521f8)) is breaking at the transformer block.

In the 560ca2f70a215e17cb6f98d52cd13efb8368a412 commit, it seems to working just fine. What should be done in this case?

fchollet:
I took a look and it seems to be failing independent of the Keras Core version?

yash-srivastava19:
Sorry for the delayed response. 

In the latest commit as well, this is working for me without any issues. I'll link the colab notebook as well, because it will be much easier to get to the issue.

https://colab.research.google.com/drive/1WzayJrt9lQIpwaSgRkVhdKcI9730bSbP?usp=sharing

If there is some issue, I'll get to it as soon as possible.

fchollet:
Keras Core is becoming Keras 3, and we're switching development to the main repository! If it is still relevant, please reopen this PR in the [keras-team/keras](https://github.com/keras-team/keras) repository. Unfortunately we aren't able to automatically transfer PRs (but we have transferred all issues).

PR #530, opened 2023-07-18
Node Test

kamathis4:


kamathis4:
Hey @fchollet, 
getting an issue with 
`import msvcrt
ModuleNotFoundError: No module named 'msvcrt'`

But only in this particular test 🙄

PR #531, opened 2023-07-18
Add Adam optimizer parallel implementation for torch backend

haifeng-jin:
With the new implementation, the `numerical_test.py` passed.


fchollet (approved):
LGTM, thank you!

PR #532, opened 2023-07-18
Add AdamW implementation with torch parallel ops

haifeng-jin:
Also changed the original AdamW to extend Adam to remove code duplication.


fchollet (approved):
LGTM, thank you

PR #534, opened 2023-07-18
Add Adadelta for torch

haifeng-jin:
The PR also contains minor bug fixes for SGD and RMSprop when using momentum.


fchollet (approved):
LGTM, thanks!

PR #540, opened 2023-07-19
Random rotations Port

kamathis4:
Trying to port Random Rotations as per this [issue](https://github.com/keras-team/keras-core/issues/481)


fchollet (commented):
You should be able to implement this using the new `affine_transform` operation.

PR #551, opened 2023-07-19
Add Nadam for torch backend

haifeng-jin:
Fixed a bug in Nadam.
`numerical_test` passed for TF, JAX, and Torch backends (need to set the `tf.keras` model to `run_eagerly=True` due to a bug in `tf.keras`).


fchollet (approved):
LGTM, thanks!

PR #552, opened 2023-07-19
Fix multi-dimension sample weights passed to mean metrics

mattdangerw:
I'm still not sure what broke this, but it appeared to change with the 0.1.1 release.

Please help me check if the outputs here are what we would expect.


mattdangerw:
OK, got this working, but found yet another error. torch is handle `axis` arguments to reductions differently then jax/tf/numpy. Can follow up with a more through change for all the reduction ops.

fchollet (approved):
Excellent, thank you!

PR #567, opened 2023-07-20
test and move backend agnostic example

anas-rz:
These examples have been tested for others backends. Should be place under backend agnostic examples.


fchollet (approved):
LGTM, thanks

PR #569, opened 2023-07-21
Add some additional TF SavedModel tests

nkovela1:
This PR adds a few additional tests for TF SavedModel. Test coverage will be expanded in the future when Keras SavedModel is integrated properly into Keras Core.


fchollet (commented):
LGTM -- thank you!

PR #575, opened 2023-07-21
Fixes Torch-GPU Test failure on initializers/random_initializers_test.py

sampathweb:
Uses `convert_to_numpy(values)` API instead of `np.array(values)`, as it doesn't work in torch backend when the tensor is on GPU.


fchollet (approved):
LGTM, thanks!

PR #576, opened 2023-07-21
Keras tensor functions

kamathis4:
Moved old branch to here due to cla isses


fchollet (approved):
LGTM, thank you for the great contribution!

PR #580, opened 2023-07-21
set torch device when converting to tensor

sampathweb:
`keras-core`, by default puts `torch` tensors into cuda when available.  When users create tensors outside of `keras-core` and try to combine with `keras-core` tensors, it causes errors as one tensor is in cuda and other in CPU.
Before the manual created torch tensor is combined with model outputs, we need to `convert_to_tensor` to put it in the same device as model data.

For example in [example](https://github.com/keras-team/keras-io/blob/master/guides/keras_core/custom_train_step_in_torch.py), following snippet would fail when running on a machine with GPU: 

```
# Decode them to fake images
generated_images = self.generator(random_latent_vectors)

# Combine them with real images
real_images = torch.tensor(real_images)
combined_images = torch.concat([generated_images, real_images], axis=0)  # Fails since two tensors are on different devices.
```
To Fix it, we need to modify it as - 

```
# Combine them with real images
real_images = torch.tensor(real_images)
real_images = keras.ops.convert_to_tensor(real_images)  # Required before combining with model outputs.
combined_images = torch.concat([generated_images, real_images], axis=0) 
```

### DIFF
--- a/keras_core/backend/torch/core.py
+++ b/keras_core/backend/torch/core.py
@@ -111,8 +111,8 @@ class Variable(KerasVariable):
 def convert_to_tensor(x, dtype=None):
     if is_tensor(x):
         if dtype is None:
-            return x
-        return x.to(to_torch_dtype(dtype))
+            return x.to(device=get_device())
+        return x.to(dtype=to_torch_dtype(dtype), device=get_device())
     if isinstance(x, Variable):
         # TorchDynamo has bugs supporting nn.Parameter type check.
         # Return it directly instead of pass it to the rest of the logic in the


haifeng-jin (commented):
[notebook link](https://colab.research.google.com/drive/1e5jrXEWBF_JrxbJ4W_YK5yxQ51vjiHxe?resourcekey=0-9RPMLjPktaE1Si4SQr-m9Q&usp=sharing)
Would you please help test the performance differences before and after this change with the code above?

`convert_to_tensor()` is heavily called with torch backend. `get_device()` is time consuming. That's why we try to remove the `.to(device)` call and try to have all the tensor's created on the device instead of use `convert_to_tensor` to move them.

For the use case you provided,
would it be possible to do `torch.tensor(real_images, device="cuda" if torch.cuda.is_available() else "cpu")`?

sampathweb:
> would it be possible to do torch.tensor(real_images, device="cuda" if torch.cuda.is_available() else "cpu")?

Sounds good.  I will use this format, so we don't pay the performance.  Will close this PR for now.

haifeng-jin:
Thanks, @sampathweb !

PR #582, opened 2023-07-22
doc: DocString for numpy.amax

pranavvp16:
Do i also need to add the shape parameter in output of the examples. I tried running it in google collab but it doesn't show shape in output


fchollet (approved):
LGTM, thanks!

PR #591, opened 2023-07-23
Update Tests for specific backend imports

sampathweb:
Allows users to not have `jax` or `torch` package unless they are using that specific backend.


fchollet (approved):
LGTM, thanks!

PR #595, opened 2023-07-24
Always install the correct version with pip_build

mattdangerw:
pip_build while create whl files in `/dist`, which aren't cleaned up by the script unless you do so manually.

The script would install the first `.whl` it saw in `/dist`, which means if you have been building multiple versions of keras-core, the install script might grab the wrong one.

This makes sure to always grab the correct versioned `.whl`. Alternately we could clear the `/dist` dir each time we build.

### DIFF
--- a/pip_build.py
+++ b/pip_build.py
@@ -81,7 +81,7 @@ def build():
 
         # Find the .whl file path
         for fname in os.listdir(dist_directory):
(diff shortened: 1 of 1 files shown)


fchollet (approved):
Thanks for the fix! I think the same script in KerasNLP might have the same issue.

PR #596, opened 2023-07-24
Fix more spots torch != numpy for reductions

mattdangerw:
This is more consistency follows up to the changes we started marking on #552.


fchollet (approved):
Neat, thank you for the fixes!

PR #600, opened 2023-07-25
Convert pretrained word embeddings to Keras Core

freedomtan:
convert Keras IO [Using pre-trained word embeddings](https://keras.io/examples/nlp/pretrained_word_embeddings) example.

1. replace tensorflow.keras with keras_core
2. don't import tensorflow, only tf.data is used
3. keras_core.initializers.Constant only accepts scalar. use Embedding.set_weights() to initialized weights

Because tf_data is used by the vectorizer, the end-to-end model in the end of the script is not backend-agnostic.

diff: 79686c3
colab [notebook](https://colab.research.google.com/drive/1AVouUHGzCTg0nsaiq_HN6U4t5fPtvAHb?usp=sharing)


fchollet (approved):
LGTM, thank you!

PR #607, opened 2023-07-25
Porting over changes to Torch and JAX keras-core distribution guides …

hertschuh:
…from keras.io.

* Most changes are reformatting.
* Some changes are to support running in Colab.


fchollet (approved):
LGTM, thanks!

PR #614, opened 2023-07-26
Port digit addition rnn example to keras-core

anas-rz:
Porting addition [example using RNN](https://keras.io/examples/nlp/addition_rnn/) to keras-core.
The model works on all backends with CPU.
On GPU it shows an error with TensorFlow which is likely from tensorflow [Colab to reproduce](https://colab.research.google.com/drive/1rBtwDh7XfYuoZVpQhs36TDZwPioid8AA?usp=sharing). 
Tested it in Colab and local environment on Ubuntu.


fchollet (approved):
LGTM -- thanks for the PR and the bug report. I fixed the bug.

PR #646, opened 2023-07-31
Removes ref in Global Average Pool and Elu/Prelu Activations

sampathweb:
Replaces `tf.keras` with numpy for expected outputs in Global Average Pooling and Elu/Prelu tests.


fchollet (approved):
Awesome -- one step closer to Keras 3!

PR #647, opened 2023-07-31
Small fixes on `image.extract_patches`

guillaumebaquiast:
Two small fixes on `extract_patches`:

* Passing argument `size` as an int currently fails for symbolic tensor:
```
>>> x = KerasTensor([None, 20, 20, 3])
>>> keras_core.ops.image.extract_patches(x, size=5)
...
    strides = (self.size[0], self.size[1])
TypeError: 'int' object is not subscriptable
```
Fixed and added unit-test.

* Examples in docstring are failing for me, I have to convert the arrays to "float32" beforehand.
```
>>> image = np.random.random((1, 20, 20, 3)) # batch of 2 RGB images
>>> patches = keras_core.ops.image.extract_patches(image, (5, 5))
...
tensorflow.python.framework.errors_impl.InvalidArgumentError: cannot compute Conv2D as input #1(zero-based) was expected to be a double tensor but is a float tensor [Op:Conv2D] name:
```
I edited the docstring accordingly.

### DIFF
--- a/keras_core/ops/image.py
+++ b/keras_core/ops/image.py
@@ -271,6 +271,8 @@ class ExtractPatches(Operation):
         data_format="channels_last",
     ):
         super().__init__()
+        if isinstance(size, int):
+            size = (size, size)
         self.size = size
         self.strides = strides
         self.dilation_rate = dilation_rate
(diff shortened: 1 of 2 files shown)


fchollet (approved):
LGTM, thanks for the fix!

PR #654, opened 2023-08-01
Performance improvement for MHA

haifeng-jin:
With this change it slightly improve the performance of BERT with torch backend.


fchollet (approved):
Thank you -- LGTM

PR #657, opened 2023-08-01
Removes `tf.keras` from Pooling tests - Max, Average and Global pooling

sampathweb:
Replaces `tf.keras` with numpy expected results for pooling layers.


fchollet (approved):
LGTM -- great work!

PR #660, opened 2023-08-02
Add a segment_max op

ianstenbit:
This op is a sibling of `segment_sum` which is useful in box decoding for a handful of 2D and 3D object detection models (and I would imagine instance segmentation models as well, although I have yet to encounter that particular use case)


fchollet (approved):
Excellent work -- looking good!

PR #662, opened 2023-08-03
Maintain keras_core as keras naming convention

grasskin:
Switched the only instance of `import keras_core` to `import keras_core as keras`. Switched original keras to `tf_keras`. Reduce namex execution paths. Other instances are all tests.


grasskin:
Closing for now - went with alternative

PR #663, opened 2023-08-03
Port `mlp_image_classification.py` to all backends

guillaumebaquiast:
Port mlp_image_classification to keras_core.

This implementation works with Tensorflow, Torch, and JAX (at least on my machine).

<ins>A few points below</ins>:

**Reimplemented positional embedding**
Just like in (PR https://github.com/keras-team/keras-core/pull/602), I ran into the issue that the example wasn’t running with a Torch backend (I had the same cryptic error message: `RuntimeError: Trying to backward through the graph a second time (or directly access saved tensors after they have already been freed). Saved intermediate values of the graph are freed when you call .backward() or autograd.grad(). Specify retain_graph=True if you need to backward through the graph a second time or if you need to access saved tensors after calling backward.`).

To fix that, I had to re-implement the logic of positional embedding.
I *believe* that the original implementation of the positional embedding made Keras ignore the embedding layer, which didn’t fail with Tensorflow or JAX (although it didn't behave as expected), but did fail for Torch. Indeed, see the difference in the FNet model summaries between the two implementations:
* With the original implementation, the embedding layer is missing:
image

* With the implementation from this PR, the embedding layer is present, and the code runs without error on a Torch backend:
image

If my hypothesis is indeed correct, then there a few other examples should get the same fix (e.g., `cct.py`, `image_captioning.py`, `token_learner.py`, `video_transformer.py`), and it should solve the issue https://github.com/keras-team/keras-core/issues/566.

**Reimplemented Patches**
This is to leverage the newly implemented `keras_core.ops.image.extract_patches`.

**Fixed some typos**
Fixed minor typos, like arguments of function that weren’t used anywhere.

**File location**
I left the file at the location `keras_io/tensorflow/vision/mlp_image_classification` for now, so that the diff displays nicely. I can move it under `keras_io/vision` before merging.

### DIFF
--- a/examples/keras_io/tensorflow/vision/mlp_image_classification.py
+++ b/examples/keras_io/tensorflow/vision/mlp_image_classification.py
@@ -1,8 +1,9 @@
 """
 Title: Image classification with modern MLP models
 Author: [Khalid Salama](https://www.linkedin.com/in/khalid-salama-24403144/)
+Converted to Keras Core by: [Guillaume Baquiast](https://www.linkedin.com/in/guillaume-baquiast-478965ba/), [divyasreepat](https://github.com/divyashreepathihalli)
 Date created: 2021/05/30
-Last modified: 2021/05/30
+Last modified: 2023/08/03
 Description: Implementing the MLP-Mixer, FNet, and gMLP models for CIFAR-100 image classification.
 Accelerator: GPU
 """
@@ -28,10 +29,8 @@ main building blocks.
 """
 
 import numpy as np
-import tensorflow as tf
 import keras_core as keras
 from keras_core import layers
-from keras_core.layers import Lambda
 
 """
 ## Prepare the data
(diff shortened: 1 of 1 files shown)


examples/keras_io/tensorflow/vision/mlp_image_classification.py
```diff
@@ -76,15 +75,11 @@ def build_classifier(blocks, positional_encoding=False):
...
+    patches = Patches(patch_size)(augmented)
     # Encode patches to generate a [batch_size, num_patches, embedding_dim] tensor.
     x = layers.Dense(units=embedding_dim)(patches)
     if positional_encoding:
-        positions = tf.range(start=0, limit=num_patches, delta=1)
-        position_embedding = layers.Embedding(
-            input_dim=num_patches, output_dim=embedding_dim
-        )(positions)
-        x = x + position_embedding
+        x = x + PositionEmbedding(sequence_length=num_patches)(x)
```
guillaumebaquiast, line 82: This change fixed the run on Torch backend, and made the embedding appear in the summary of the model.

examples/keras_io/tensorflow/vision/mlp_image_classification.py
```diff
@@ -168,26 +163,74 @@ def run_experiment(model):
...
+        num_patches = keras.ops.shape(patches)[1] * keras.ops.shape(patches)[2]
+        patch_dim = keras.ops.shape(patches)[3]
+        out = keras.ops.reshape(patches, (batch_size, num_patches, patch_dim))
+        return out
+
+"""
+## Implement position embedding as a layer
+"""
+
+class PositionEmbedding(keras.layers.Layer):
```
guillaumebaquiast, line 182: This class was adapted from KerasNLP.

fchollet (approved):
The fix makes sense to me -- thank you for debugging this! LGTM

PR #666, opened 2023-08-04
update requirements for all cpu install

haifeng-jin:
Minor corrections for `requirements.txt` to use all CPU requirements.
To use GPUs, one should install `requirements-cuda.txt`.


fchollet (approved):
LGTM, thanks!

PR #668, opened 2023-08-06
Backend agnostic port of object detection using vit

soumik12345:
Backend-agnostic port of the example [Object detection with Vision Transformers](https://keras.io/examples/vision/object_detection_using_vision_transformer) that works with TF, Jax, and Torch backends. Based on the TF-compatible port originally contributed by @grasskin as part of https://github.com/keras-team/keras-core/pull/313.

Colab Notebook: https://colab.research.google.com/drive/1-HUon9woz5yag7rVtOYbsLnwPX3pbZCW?usp=sharing


fchollet (approved):
Excellent work! LGTM

PR #669, opened 2023-08-06
Port Compact Convolutional Transformers to backend agnostic

guillaumebaquiast:
Port Compact Convolutional Transformers to be backend agnostic. Works on TF, Jax, and Torch backends (at least on my machine).

Two notes:
* I used the same fix as in https://github.com/keras-team/keras-core/pull/663 to make the Position Embedding work (without the fix, it fails with Torch, and doesn't behave as expected on Jax and Tensorflow)
* The implementation of `StochasticDepth` was failing for Torch, due to what I believe was a typo from the original implementation.


examples/keras_io/tensorflow/vision/cct.py
```diff
@@ -144,32 +139,77 @@ def call(self, images):
...
-            embed_layer = layers.Embedding(
-                input_dim=sequence_length, output_dim=projection_dim
+
+"""
+Positional embeddings are optional in CCT. If we want to use them, we can use
+the Layer defined below.
+"""
+
+
+class PositionEmbedding(keras.layers.Layer):
```
guillaumebaquiast, line 159: Same as in https://github.com/keras-team/keras-core/pull/663

examples/keras_io/tensorflow/vision/cct.py
```diff
@@ -214,9 +255,9 @@ def __init__(self, drop_prop, **kwargs):
     def call(self, x, training=None):
         if training:
             keep_prob = 1 - self.drop_prob
-            shape = (tf.shape(x)[0],) + (1,) * (tf.shape(x).shape[0] - 1)
-            random_tensor = keep_prob + tf.random.uniform(shape, 0, 1)
-            random_tensor = tf.floor(random_tensor)
+            shape = (keras.ops.shape(x)[0],) + (1,) * (len(x.shape) - 1)
```
guillaumebaquiast, line 258: `(tf.shape(x).shape[0] - 1)` made Torch fail (error message was complaining about too many dimensions to the tensor). I believe it was a typo from the original implementation, which I replaced by `(len(x.shape) - 1)`

fchollet (commented):
Thanks for the PR! The changes look good. Since the example is backend-agnostic, you can move it outside of the `tensorflow` folder (to `keras_io/vision`).

fchollet (approved):
LGTM, thank you!

PR #672, opened 2023-08-07
Porting text Preprocessing from tf.keras to keras_core

asingh9530:
Hi @fchollet ,

This PR is specifically for porting Text-Preprocessing layer from tf.keras to keras_core.

Note not yet completed.

Issue : https://github.com/keras-team/keras-core/issues/667

Thanks

### DIFF
--- a/keras_core/layers/preprocessing/text_vectorization.py
+++ b/keras_core/layers/preprocessing/text_vectorization.py
@@ -6,7 +6,33 @@ from keras_core.layers.layer import Layer
 from keras_core.saving import serialization_lib
 from keras_core.utils import backend_utils
 from keras_core.utils.module_utils import tensorflow as tf
+from keras_core.layers.preprocessing.string_lookup import StringLookup
+from keras_core.utils import layer_utils
 
+
+keras_kpl_gauge = tf.__internal__.monitoring.BoolGauge(
+    "/keras_core/layers/preprocessing",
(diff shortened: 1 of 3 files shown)


asingh9530:
Hi @fchollet,

Need confirmation from you following needs to be ported.

- `DataHandler` as adapt will require it, so need your confirmation since in `keras_core` we don't have engine how about we use utils instead.
-  Also since `textVectorization` depends on `StringLookup` so it is needed to migrate `StringLookup` also in this PR.

asingh9530:
> Thanks for the PR. I took a close look at the contents, and unfortunately there's a huge amount of things that would need to be removed or refactored. The problem is that this is not meant to be a 1:1 copy of the code in tf.keras. There's a significant amount of adaption that's needed. E.g. ` tf.__internal__.monitoring.BoolGauge` is not needed, references to `tf.compat.v1` need to be removed, error messages must be adapted to refer to Keras Core, etc.
> 
> I appreciate your efforts to make a contribution, but I'm afraid this item is not turning out to be a good match. You could take on another item instead: #481

Hi @fchollet, Yes I agree this PR will require a big change and refactoring. I wanted to know do you have any RFC or document for `keras_core` as it will be easier to make large changes which might require structural changes or large refactoring. If it is something internal then its fine otherwise it will be very helpful. 🙏🏻

PR #674, opened 2023-08-07
Change saving API routing to deprecate SavedModel format

nkovela1:
This PR changes the saving API to deprecate the SavedModel format (which will not be included in Keras Core due to backwards incompatibilities). This includes a recommendation for `tf.saved_model.save()` for export.

Additionally, this means the `save_format` flag will be unnecessary and users should either pass a `.h5` or `.keras` extension for file paths. I've included a warning where the flag is redundant (i.e. when the user passes a valid extension and then also passes the `save_format` flag).

Associated tests have also been added to test for these API errors as well.


fchollet (approved):
LGTM, thank you!

PR #675, opened 2023-08-08
Make ner transformers work again

freedomtan:
1. 489237f  breaks this example,  don't use tensor in MHA `__init__()`, ed2a726
2. keras.losses.Reduction.NONE  is no longer there, replace it with `None`


fchollet (approved):
LGTM, thanks

PR #687, opened 2023-08-10
Convert to tensor for jax ops.numpy.transpose

mattdangerw:
Another spot we need this when passing a variable.


fchollet (approved):
LGTM, thanks!

PR #698, opened 2023-08-11
Random Flip for batched inputs

guillaumebaquiast:
Random Flip seems to fail for batched inputs.

For instance:
```
>>> (x_train, y_train), (x_test, y_test) = keras_core.datasets.cifar10.load_data()
>>> keras_core.layers.RandomFlip(mode="horizontal")(x_train[1]) # Works
>>> keras_core.layers.RandomFlip(mode="horizontal")(x_train) # Fails
InvalidArgumentError                      Traceback (most recent call last)
...
{{function_node __wrapped__SelectV2_device_/job:localhost/replica:0/task:0/device:CPU:0}} condition [50000], then [50000,32,32,3], and else [50000,32,32,3] must be broadcastable [Op:SelectV2] name:
```

This is due to some shape broadcasting gone wrong, which this PR should fix. Added unit-test as well.


fchollet (approved):
LGTM -- thanks for the fix!

PR #705, opened 2023-08-12
Add activation docstrings // ops.nn.py

Frightera:
This PR completes the docstrings of activations available under `ops.nn.py`


fchollet (approved):
Thank you for the PR! LGTM

PR #706, opened 2023-08-12
Remove tf dependency from math_test

Frightera:
This PR removes tensorflow operations and hard-codes the expected values when possible.

It's not `tf.keras` and was not mentioned in the pinned request, but might be something useful?


fchollet (approved):
LGTM, thank you!

PR #713, opened 2023-08-13
Fix Issue in SegmentMax's Symbolic Call in keras_core/ops/math.py

Faisal-Alsrheed:
**Background**
The current implementation of the segment_max function in keras_core.ops.math has an error when dealing with symbolic tensors. Specifically, when checking for symbolic tensors and attempting to use the SegmentMax class's symbolic call method, it erroneously calls SegmentSum instead of SegmentMax.

**Changes**
Updated the segment_max function's symbolic tensor handling to call the correct SegmentMax class's symbolic_call method.

**Impact:**
Fixing this ensures that the segment_max function behaves correctly for symbolic tensors and avoids incorrect computations


fchollet (approved):
LGTM, thanks for the fix!

PR #721, opened 2023-08-14
Only compute mean once in layernorm

mattdangerw:
Small think I just noticed, the `tf.keras` layernorm was using `tf.nn.moments` to compute mean and variance together.

However, we are computing them separately with numpy here, meaning that we were actually computing the mean of the inputs twice. Should be more efficient to reuse the computed mean to compute the variance.


mattdangerw:
Test failure looks unrelated.

fchollet (approved):
LGTM

PR #722, opened 2023-08-14
Always return a tuple from ops.shape

mattdangerw:
This was already true in all contexts except when called on `tf.Tensor`s on the Tensorflow backend. In this case, we were returning a 1-D `tf.Tensor`.

This was leading to some weird incongruencies, e.g.
```python
# On Jax.
ops.shape(np.ones((1, 2))) + (3,)
> (1, 2, 3)

# On TensorFlow.
ops.shape(np.ones((1, 2))) + (3,)
> tf.Tensor([4, 5])
```

### DIFF
--- a/keras_core/backend/tensorflow/core.py
+++ b/keras_core/backend/tensorflow/core.py
@@ -87,7 +87,23 @@ def is_tensor(x):
 
 
 def shape(x):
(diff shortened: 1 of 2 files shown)


fchollet (approved):
LGTM, thank you

PR #724, opened 2023-08-14
Add some additional H5 saving test coverage

nkovela1:
This PR adds some additional test coverage for H5 with Keras Core.

In the future:
After all PRs are finished for the H5 format testing, I will generate the models in TF and their goldens as files to prevent dependency on TF for this testing suite.


fchollet (approved):
LGTM, thank you!

PR #725, opened 2023-08-14
Replace `tf.keras` for expected result in Conv1D Test

sampathweb:
Replaces `tf.keras` with numpy implementation for `conv1d` test.  Will do the same for Conv2D and Conv3D in a subsequent PR.


fchollet (approved):
LGTM, thank you

PR #740, opened 2023-08-16
Replace `tf.keras` with numpy  for conv2d and conv3d tests

sampathweb:
* Replaces `tf.keras` with numpy reference implementation
* Adds test for `channel_first` and changing dilation rates.
* Replaces `tf.keras` in `numerical_utils_test.py`

### DIFF
--- a/keras_core/layers/convolutional/conv_test.py
+++ b/keras_core/layers/convolutional/conv_test.py
@@ -1,8 +1,5 @@
-import math
-
 import numpy as np
 import pytest
-import tensorflow as tf
 from absl.testing import parameterized
 from numpy.lib.stride_tricks import as_strided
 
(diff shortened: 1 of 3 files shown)


fchollet (commented):
Thanks for the PR! Some tests in JAX are failing -- it seems the precision of certain numerical checks needs to be reduced to 1e-5.

sampathweb:
> Thanks for the PR! Some tests in JAX are failing -- it seems the precision of certain numerical checks needs to be reduced to 1e-5.

Yes, it fails sometimes on `conv_transpose_test` - I reduced diff to 1e-5.  I also plan to send a PR and remove `tf.keras` from `conv_transpose_test` before end of this week.

fchollet (approved):
LGTM

PR #747, opened 2023-08-18
Fixed uniqueness bug and bits->bytes

jackd:
This fixes two issues:
1. The original unique-ifying code did nothing. This is fixed.
2. The return values are now in bytes, consistent with the documentation and usage elsewhere in keras-core.


fchollet (approved):
Indeed -- thanks for the fix!

PR #752, opened 2023-08-19
Remove `tf.keras` from depthwise and separable conv tests

sampathweb:
* Removes `tf.keras` from Depthwise and Separable conv tests.
* Implemented separable conv as a combination of depthwise followed by regular conv with 1x1 kernel.  So imported the numpy test implementations for separable conv test.

Will send another PR to remove `tf.keras` from Transpose Conv test.


fchollet (approved):
LGTM -- thank you!

Issue #719, opened 2023-08-14
How to get PyTorch Module from Keras Core model?


goto-loop:
I would like to compile a Keras Core model (PyTorch backend) to Apache TVM following [this guide](https://tvm.apache.org/docs/how_to/compile_models/from_pytorch.html). Part of the conversion is tracing the model:
``` python
input_data = torch.randn(input_shape)
scripted_model = torch.jit.trace(model, input_data).eval()
```
This doesn't seem to work when `model` is a Keras Core model, so I would like to convert it to a PyTorch `Module` first. The [Keras Core announcement](https://keras.io/keras_core/announcement/) says "Any Keras Core model can be instantiated as a PyTorch Module", so I assume this is possible – but how to go about it?

haifeng-jin:
Hi @goto-loop ,
Any Keras Core model is a PyTorch `nn.Module` subclass if you use PyTorch backend.
However, we are having issues in supporting any `torch.jit` functions or `torch.compile` right now.
They can now only run in eager mode.

james77777778:
Hi @haifeng-jin 
After making a small fix to `keras_core/backend/common/variables.py`, I managed to do the following:

```python
import os

os.environ["KERAS_BACKEND"] = "torch"

import torch

from keras_core.applications.mobilenet_v3 import MobileNetV3Small

input_shape = (224, 224, 3)
input_data = torch.randn(size=(1, *input_shape))
model = MobileNetV3Small(input_shape=input_shape, minimalistic=True)

# torch.compile
compiled_model = torch.compile(model)
_ = compiled_model(input_data)  # might take some time...
print("finished compiling!")

# torch.jit
scripted_model = torch.jit.trace(model, input_data).eval()
torch.jit.save(scripted_model, "mobilenetv3small.pt")
print("finished tracing!")

# torch.onnx
torch.onnx.export(scripted_model, input_data, "mobilenetv3small.onnx")
print("finished onnx exporting!")

```

Although the exported model was full of weired operators (possibly due to using `channel_last`), `torch.compile`, `torch.jit.trace` and `torch.onnx.export` work fine.

The fix is in `standardize_shape`:

```python
def standardize_shape(
    shape, allow_dynamic_batch_size=True, allow_all_dynamic=True
):
    if not isinstance(shape, tuple):  # -> this does not work well when using torch
        if shape is None:
            raise ValueError("Undefined shapes are not supported.")
        if not hasattr(shape, "__iter__"):
[...]

PR #753, opened 2023-08-20
Fix `standardize_shape`

james77777778:
As discussed in #719 

`torch.Size` is a subclass of tuple, so it might break the logic in `standardize_shape` that `isinstance(e, int)` is invalid for `torch.Tensor`.

cc @haifeng-jin


fchollet (approved):
LGTM, thanks

PR #758, opened 2023-08-21
Enhance docstrings + easy to understand examples (nn.py)

Faisal-Alsrheed:
As promised, I have begun enhancing the docstrings and adding easy-to-understand examples. 
The goal is to make the Keras core documentation a pleasant resource for learning deep learning. 
The objective of these straightforward examples is to facilitate understanding without delving into complex mathematics.


fchollet (approved):
LGTM, thank you

PR #759, opened 2023-08-21
Fix naming in `mobilenet_v3` and `densenet`

james77777778:
This PR fixes the naming error in MobileNetV3 and DenseNet caused by the introduction of `name_scope`.

A new assertion in `keras_core/ops/operation.py` and corresponding tests have been added.


fchollet (approved):
LGTM, thank you for the fix!

PR #773, opened 2023-08-22
Remove dependencies on tf.keras in nn_test.py

Faisal-Alsrheed:
I have started to remove dependencies on tensorflow in nn_test.py.

Please share your feedback before doing the rest.


fchollet (commented):
Thanks for the PR! Mind you, this is a dependency on `tf.nn` rather than `tf.keras`. We don't need to get rid of `tf` usage (outside `tf.keras`) because it's generally fine to have our unit tests depend on TF (or JAX, or numpy, or torch). It's not fine for `tf.keras` because that would become self-referential in the future. We're only looking to eliminate `tf.keras` in unit tests specifically.

PR #775, opened 2023-08-22
Add more H5 backwards compatibility tests and lambda support

nkovela1:
This PR extends backwards compatibility tests for the HDF5 format and adds backwards support for Lambdas.


fchollet (approved):
LGTM, thank you

PR #782, opened 2023-08-23
Removes `tf.keras` from transposed conv test

sampathweb:
Removes `tf.keras` dependency from transposed conv test and replaces it with numpy expected results implementation.


fchollet (approved):
LGTM, thank you. What are the tf.keras dependencies remaining at this time?

sampathweb:
> LGTM, thank you. What are the tf.keras dependencies remaining at this time?

Its primarily in preprocessing tests.  I plan to take on that next with ETA of early next week.

PR #787, opened 2023-08-24
Replace `tf.data` with `tf_data` in preprocessing tests

sampathweb:
Replaces `tf.data` usages with `from tensorflow import data as tf_data` in preprocessing tests.


fchollet (approved):
LGTM

PR #788, opened 2023-08-24
fixed squeeze axis inconsistency

jackd:
The `Squeeze` op is confused as to whether it should accept `axis` in the constructor or as an argument in `call` / `compute_output_spec`. This makes it self-consistent, and what I think is consistent with the other ops.


fchollet (approved):
LGTM, thanks

PR #789, opened 2023-08-24
Split fix

jackd:
Split's `compute_output_spec` does not return the final element (though the underlying `call` does). This addresses this, plus allows passing tuples to `keras.ops.split` without error.


fchollet (commented):
Thanks for the PR. Please add a unit test.

fchollet (approved):
LGTM, thank you!

PR #790, opened 2023-08-25
Fix argument order in stft and istft

james77777778:
(Sorry for the typo in previous PR)

The argument order in the stft and istft functions was incorrect.
`(sequence_length, sequence_stride, fft_length, center, window)` should be `(sequence_length, sequence_stride, fft_length, window, center)`

This PR addresses it by using kwargs to validate the provided args.


AakashKumarNain (approved):

fchollet (approved):
LGTM

PR #792, opened 2023-08-25
Fix channels_first format bug for rescaling layer

haifeng-jin:
The `EfficientNet` in Keras applications uses the Rescaling layer with a vector (`shape=(channels,)`) as the scale.
See https://github.com/keras-team/keras-core/blob/v0.1.5/keras_core/applications/efficientnet.py#L333-L335


fchollet (approved):
LGTM, thanks for the fix!

PR #795, opened 2023-08-26
Update example losses on docstrings

Frightera:
It is better to demonstrate those metrics using suitable loss functions rather than `mse`


fchollet (approved):
LGTM

PR #799, opened 2023-08-27
Increase-test-coverage-in-activations_test.py

Faisal-Alsrheed:
This PR introduces improved tests for several activation functions, ensuring their robustness and precision:

These enhancements ensure that the activation functions behave as expected across various scenarios, contributing to the reliability and stability of the Keras-core.

`Leaky ReLU`: Added tests for various negative_slope values.
`ReLU6`: Enhanced tests to cover edge cases.
`SiLU/Swish`: Introduced tests for the activation function.
`GELU`: Added separate tests for approximate values, both True and False.
`Hard Sigmoid`: Extended the test to include more scenarios, ensuring better coverage.
`Sigmoid`: Improved test to cover a wider range of input values.
`SoftSign`: Adjusted the test for increased reliability.
`Softplus`: Extended the existing test for better coverage.
`Exponential`: Refined the test to capture more diverse cases.
`Tanh`: Expanded the test to include a broader range of values.

### DIFF
--- a/keras_core/activations/activations_test.py
+++ b/keras_core/activations/activations_test.py
@@ -323,46 +323,14 @@ class ActivationsTest(testing.TestCase):
         expected_3d = np.vectorize(_ref_hard_sigmoid)(x_3d)
         self.assertAllClose(result_3d, expected_3d, rtol=1e-05)
 
-        # Test with strictly positive values greater than 1
-        x_positive_above_1 = np.random.uniform(1.1, 10, (2, 5))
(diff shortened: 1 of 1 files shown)


fchollet (approved):
Thanks for the PR! It looks good -- but generally the uselessness of testing more *input values* is very limited. We should aim at *code coverage* first, as in, making sure that every line in the codebase is run at least once in a unit test. That is the most useful kind of unit test we can add right now -- tests that increase code coverage.

PR #802, opened 2023-08-27
fixed apply_mask bug in losses

jackd:
This fix ensures that losses evaluated with a `sample_weights == ops.cast(mask, "float32")` are the same as those evaluated with either argument but not both `None`. Note this `total` is consistent with the `batch_size` used in `reduce_values`, corresponding to the total size of all elements as opposed to the leading dimension. Another fix might be to leave `apply_mask` alone and change `reduce_values` to divide by the leading dimension only rather than the total size.

I'm unsure of a concise way to unit test this because AFAIK there's no way to pass a mask in using only the public API...


fchollet (commented):
Thanks for the fix -- please add a unit test for this case.

fchollet (approved):
LGTM, thank you! 👍

PR #804, opened 2023-08-28
Replace `tf.keras` in center-crop and resize tests

sampathweb:
Replaces `tf.keras` in Center Crop and Resizing pre-processing layers.

This completes the task of removing `tf.keras` in all test reference outputs.  Its only used in `legacy/saving/legacy_h5_format.test`


fchollet (approved):
LGTM, thank you!

For the saving tests, @nkovela1 will remove the reliance on tf.keras by using externally loaded saved model files.

PR #815, opened 2023-08-30
Don't double cast layer norm weights in mixed precision

mattdangerw:
Previously, under mixed precision, we would autocast full precision variables to half precision, then manually cast back to full precision.

Possibly this would compile away (no idea), but is certainly very slow in eagerly. We should set `autocast=False` to avoid the needless type conversion loop.


fchollet (approved):
LGTM, thanks

PR #817, opened 2023-08-30
Fix the Jax distributed variable load issue.

qlzh727:
The jax varaible will preserve the layout information at the init time, and will reuse it for any future assigned the value.

This will fix the issue for keras nlp/cv model that does the weight loading after the model creation, which unshard the variable back to the device:0


fchollet (approved):
LGTM, thanks. Does this mean that model.save()/load() now works with sharded variables?

qlzh727:
> LGTM, thanks. Does this mean that model.save()/load() now works with sharded variables?

The model.load() should work with the sharded variable now. The saving behavior is not changed by this PR.

PR #819, opened 2023-08-30
Add Export for JAX Backend

nkovela1:
This PR adds initial support for the export of JAX-backend models to TF SavedModel.
Additionally, a test has been added for subclassed model export support along with logic in both TF and JAX backends.

There are a few caveats and TODOs to note here:

- jax2tf function conversion is lacking argument name preservation, causing the names of arguments passed to become `args_tf_0`, `args_tf_1`, etc. I have opened a bug internally in JAX2TF Users regarding this, and a workaround should be included in a subsequent PR.

- ReloadedLayer support will be included in a subsequent PR, further discussion will be needed on this.

- All other tests pass!


keras_core/backend/common/variables.py
```diff
@@ -432,6 +432,9 @@ def standardize_shape(shape):
     for e in shape:
         if e is None:
             continue
+        if config.backend() == "jax" and str(e) == "b":
+            # JAX2TF tracing represents `None` dimensions as `b`
```
fchollet, line 436: Are they literally the string `"b"`? Is there a more reliable way to make this check than `str(e) == "b"`?

keras_core/layers/layer.py
```diff
@@ -1217,6 +1221,32 @@ def _get_call_context(self):
             self._clear_losses()
         return layer_call_ctx
 
+    def _set_save_spec(self, args, kwargs):
```
fchollet, line 1224: This is currently causing some layer tests to fail. It's also partially redundant with another feature that we have, which is that we record the shapes of all tensor input arguments during `build()`.

You should be able to remove this and instead move the logic to the exporter. You can access `_build_shapes_dict` for the shapes of all input arguments (in dict format, e.g. with keys like `"input_shape"`). The dtypes are normally going to be `self.input_dtype`.

Something like:

```python
if len(shapes_dict) == 1:
   return TensorSpec(shape=shapes_dict.values()[0], dtype=self.input_dtype)
specs = {}
for key, value in shapes_dict.items():
   key = removesuffix(key, "_shape")
   specs[key] = TensorSpec(shape=value, dtype=self.input_dtype)
return specs
```

fchollet (commented):
Thanks for the PR!

keras_core/export/export_lib.py
```diff
@@ -81,32 +83,65 @@ def __init__(self):
...
-        self._tf_trackable.variables = []
-        self._tf_trackable.trainable_variables = []
-        self._tf_trackable.non_trainable_variables = []
+        if backend.backend() == "tensorflow":
+            self._tf_trackable = tf.__internal__.tracking.AutoTrackable()
+            self._tf_trackable.variables = []
+            self._tf_trackable.trainable_variables = []
+            self._tf_trackable.non_trainable_variables = []
+        else:  # JAX backend
+            self.module = tf.Module()
```
fchollet, line 92: Use a private attribute -- also is this the right class? Why is AutoTrackable inappropriate for JAX?

keras_core/export/export_lib.py
```diff
@@ -81,32 +83,65 @@ def __init__(self):
...
+                "The export API is only compatible with JAX and TF backends."
+            )
 
     @property
     def variables(self):
-        return self._tf_trackable.variables
+        if backend.backend() == "tensorflow":
+            return self._tf_trackable.variables
+        else:  # JAX backend
+            return self._variables
```
fchollet, line 108: There's too much divergence between the JAX and TF paths. Either we find a way to unify (use `_tf_trackable`), or we should have separate subclasses for each backend.

keras_core/layers/layer.py
```diff
@@ -626,6 +627,7 @@ def compute_mask(self, inputs, previous_mask):
     @traceback_utils.filter_traceback
     def __call__(self, *args, **kwargs):
         self._check_super_called()
+        self._called = True
```
fchollet, line 630: Why not use `self.built`? Should be equivalent

keras_core/layers/layer.py
nkovela1: Great suggestion, I've fixed this according to the flow you suggested, thanks!

keras_core/layers/layer.py
nkovela1: If a user calls `build` by themselves, for example, we should not be allowing export. There is an existing test for this exact flow in TF Keras:

https://github.com/keras-team/keras/blob/master/keras/export/export_lib_test.py#L364

It's a niche case, but let me know what you think.

keras_core/export/export_lib.py
nkovela1: Good point, I've unified the code paths with _tf_trackable. Thanks!

keras_core/export/export_lib.py
nkovela1: Unified with AutoTrackable, thanks for the great suggestion!

keras_core/backend/common/variables.py
nkovela1: Yes, in JAX the dimensions unknown at tracing time are represented by the literal string "b", so this is the easiest check:

https://github.com/google/jax/blob/main/jax/experimental/jax2tf/README.md#shape-polymorphic-conversion

fchollet: What about just `e is "b"`?

keras_core/export/export_lib.py
```diff
@@ -100,13 +107,25 @@ def non_trainable_variables(self):
...
                 f"Received instead an object of type '{type(resource)}'. "
                 f"Object received: {resource}"
             )
+        if backend.backend() == "jax" and not isinstance(
+            resource, backend.jax.layer.JaxLayer
+        ):
+            raise ValueError(
+                "Invalid resource type. Expected an instance of a "
+                "JAX-based Keras `Layer` or `Model` "
+                "(backend.jax.layer.JaxLayer). "
```
fchollet, line 125: This is private, let's not mention it in the error message

keras_core/layers/layer.py
```diff
@@ -244,6 +244,7 @@ def __init__(
         self.dtype_policy = mixed_precision.resolve_policy(dtype)
         self.autocast = autocast
         self._input_spec = None
+        self._called = None
```
fchollet, line 247: False

fchollet (commented):
Thanks for the update!

keras_core/backend/common/variables.py
nkovela1: That results in a SyntaxError with Python:

image

I think it used to be a SyntaxWarning but more recently upgraded to Error.

keras_core/layers/layer.py
nkovela1: Fixed, thanks.

keras_core/export/export_lib.py
nkovela1: Removed, thanks.

fchollet (approved):
LGTM, thanks!

PR #823, opened 2023-08-30
Don't lock functional models

mattdangerw:


mattdangerw:
We could go even further than this, and allow all `keras.Model` classes to be mutable after build. Probably a similar use case, attaching a layer to the model that you want attached/saved, but that isn't directly used by `build/call` in the main forward pass.

qlzh727:
Let's add a quick unit test for this.

fchollet (approved):
Makes sense to me. Also, subclassed models should behave like layers from a state management standpoint, so no need to extend the fix to all models.

PR #825, opened 2023-08-31
Add rank>2 support for `stft`

james77777778:
After experimenting with `stft` and `istft`, I have discovered that the input rank limitation imposed by `torch.stft` is inconvenient.

When containing multiple channels in single waveform, the input shape might be `(batch_size, num_samples, num_channels)` which is not supported by `torch.stft`. However, this format works well with other backends.

To address this, the PR has added `reshape` logic for `stft` function in the torch backend. The rank limitation of `ops.stft` has now been removed.

Furthermore, I have found that `torch.istft` works under specific condition which is common in audio processing.

The current implementation can be summarized in the following table:

|backend|stft|istft|
|-|-|-|
|numpy|`scipy.signal.stft`|`scipy.signal.istft`|
|jax|`jax.scipy.signal.stft`|`jax.scipy.signal.istft`|
|tf|`tf.signal.stft`|`tf.signal.inverse_stft`|
|torch|`torch.stft`|`torch.istft` & custom implementation|

### DIFF
--- a/keras_core/backend/torch/math.py
+++ b/keras_core/backend/torch/math.py
@@ -282,6 +282,13 @@ def stft(
     else:
         win = torch.ones((sequence_length,), dtype=x.dtype, device=get_device())
 
+    need_unpack = False
+    *batch_shape, samples = x.shape
+    if len(x.shape) > 2:
+        need_unpack = True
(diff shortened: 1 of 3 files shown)


fchollet (approved):
LGTM, thank you!

PR #827, opened 2023-08-31
Fix weird broken test that never ran

mattdangerw:
Just stumbled across this, we had a test that didn't have `test_` in the name, was never run, and was very broken if you tried to run it.


fchollet (approved):
Thanks for the fix!

PR #831, opened 2023-09-01
Replace assertTrue(isinstance(...)) with assertIsInstance for easier …

hertschuh:
…debugging.

`assertIsInstance` prints out "x is not an instance of y".

`assertTrue(isinstance(...))` prints out "False is not True".


fchollet (approved):
LGTM, thank you

PR #840, opened 2023-09-06
Enable codecov

kiukchung:
Enables codecoverage collection while running tests and uploads to codecov.io


fchollet (approved):
LGTM, thanks!

PR #848, opened 2023-09-07
Update minimum python version to 3.9 to align with TF/JAX

sampathweb:
Currently the tests are running with `JAX==0.4.13` which was the last version that supported Python 3.8.  The latest `jax` version is `0.4.14`.  TensorFlow also has dropped python 3.8 in the upcoming `2.14` release.  Numpy has already dropped it.  So updating `keras-core` to use Python 3.9 as the minimum version.

### DIFF
--- a/.github/workflows/actions.yml
+++ b/.github/workflows/actions.yml
@@ -11,7 +11,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: [3.8]
+        python-version: [3.9]
         backend: [tensorflow, jax, torch, numpy]
     name: Run tests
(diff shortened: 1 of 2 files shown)


ianstenbit:
The PR title currently says 3.8, but I think it should be 3.9, right?

I think updating to 3.9 for KerasCV should be fine.

sampathweb:
> The PR title currently says 3.8, but I think it should be 3.9, right?
> 
> I think updating to 3.9 for KerasCV should be fine.

Thanks.  Updated the PR title to 3.9.

fchollet (approved):
LGTM, thanks

PR #849, opened 2023-09-07
Fix autocast scope again

mattdangerw:
Sometimes in preprocessing layers our dtype is not a floating point dtype. In these cases we should never autocast, it would error immediately.


fchollet (commented):
Thanks for the PR! Please add a test and a code comment.

mattdangerw:
Done!

fchollet (approved):
LGTM, thanks!

PR #850, opened 2023-09-07
Update return type for optimizer.apply

mattdangerw:
Style nit, remove return type from `optimizer.apply`, so it is 1-1 with the signature for `optimizer.stateless_apply`. We can leave the iterations return type on `apply_gradients` for compat with tf.keras. It seems mostly for compat, and has no stateless version.


mattdangerw:
The torch failure here looks unrelated. Could it be a flake?

fchollet (approved):
LGTM, thanks

PR #863, opened 2023-09-11
Add a note in README.md regarding pre-import of backend

### LINKED ISSUE #833: Improve Documentation on KERAS_BACKEND
I was testing a sample Kaggle competition notebook found here https://www.kaggle.com/competitions/rsna-2023-abdominal-trauma-detection/code

I wrote a simple script to load the model and ran into an issue. It took me a few days to figure out why.

if I set KERAS_BACKEND after importing keras libraries, models.load_model fails with a strange error.  Once I moved line to the top, it got rid of the error. It would be good to update the documentation so that users can resolve this quickly without wasting a bunch of time. I tend to import all the libraries first and then set any environment variables after the imports. I can make a documentation pull request if that helps.

import os
# backend must be set before loading keras_core to avoid load errors
os.environ["KERAS_BACKEND"] = "tensorflow"
import sys
import keras_cv
import keras_core as keras
from keras_core import layers

import numpy as np
import pandas as pd
import tensorflow as tf

class Config:
    SEED = 61
    IMAGE_SIZE = [256, 256]
    BATCH_SIZE = 160
    EPOCHS = 10
    TARGET_COLS  = [
        "bowel_injury", "extravasation_injury",
        "kidney_healthy", "kidney_low", "kidney_high",
        "liver_healthy", "liver_low", "liver_high",
        "spleen_healthy", "spleen_low", "spleen_high",
    ]
    BASE_PATH = "./input/rsna-2023-abdominal-trauma-detection"
    AUTOTUNE = tf.data.AUTOTUNE

config = Config()

def main():
    args = sys.argv[0:]
    if len(args) == 1:
        print('Example usage:')
        print('    python train_model.py checkpoint_path')
    else:
        print(args)
        filename = args[1]
        print("try to load the model")
        loadmodel = keras.models.load_model(filename)
        print(loadmodel.summary())

if __name__ == "__main__":
    main()

SuryanarayanaY:
With `keras_core` users has to import the **backend** first before importing `keras_core`. It has to be documented in **README.md** for better visibility and reach to the community.

Fixes #833

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,7 +49,8 @@ os.environ["KERAS_BACKEND"] = "jax"
 
 import keras_core as keras
 ```
-
+**Note:** The backend must be configured before importing keras_core, and the backend cannot be changed after 
+the package has been imported.
 ## Backwards compatibility
 
 Keras Core is intended to work as a drop-in replacement for `tf.keras` (when using the TensorFlow backend). Just take your


fchollet (approved):
LGTM, thanks

PR #865, opened 2023-09-12
Add support for Tensorflow SparseTensors: reshaping layers.

hertschuh:
Added `tf.SparseTensor` support for ops:
- reshape
- transpose

Added `tf.SparseTensor` support for reshaping layers:
- Flatten
- Permute
- Reshape


fchollet (approved):
Nice work! LGTM

PR #867, opened 2023-09-12
Don't convert int array input to floatx

mattdangerw:
In tf.keras, int array input is not converted to floatx. In keras-core, all array input is converted to flaotx.

This feels both like a regression compatibility wise, and quite surprising generally for subclass models. There are plenty of valid reasons to pass some int input to a model.

https://colab.research.google.com/gist/mattdangerw/2a030f1b14ec4fa3347711e84541e6d7/array-conversion-discrepency.ipynb

### DIFF
--- a/keras_core/trainers/data_adapters/array_data_adapter.py
+++ b/keras_core/trainers/data_adapters/array_data_adapter.py
@@ -276,7 +276,7 @@ def can_convert_arrays(arrays):
     )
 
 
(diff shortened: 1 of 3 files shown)


fchollet (commented):
Thanks for the PR! The code looks good, please increase test coverage.

mattdangerw:
Yeah, some missing cases for this class generally. Added.

fchollet (approved):
LGTM, thank you!

PR #868, opened 2023-09-12
Respect the $KERAS_HOME environment variable for get_file

mattdangerw:
Let me know if we want a test for this somehow, we don't really have test coverage for config.py or keras.json right now.


fchollet (approved):
LGTM, thank you!

PR #872, opened 2023-09-13
Fix `stop_gradient` in np

james77777778:
We should return `x` instead of `None` in numpy's `stop_gradient`.

A unit test has been added to verify the return value.


fchollet (approved):
LGTM, thank you

Issue #881, opened 2023-09-14
Run actions.yml with read-only permissions


pnacht:
The `actions.yml` workflow currently runs with `write-all` permissions. [GitHub recommends](https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions#using-secrets) running workflows with minimal permissions.

`actions.yml` seems to only run tests, and therefore doesn't need broad permissions. 

This issue can be solved in two ways:

- add top-level read-only permissions to the workflow, and/or
- set the default token permissions to read-only in the repo settings.

I'll be sending a PR along with this issue that sets the read-only permissions. If you also/instead wish to modify the default token permissions:

- Go to [Settings > Actions > General](https://github.com/keras-team/keras-core/settings/actions)
- Under "Workflow permissions", set them to "Read repository contents and packages permissions"

---

**Disclosure:** My name is Pedro and I work with Google and the [Open Source Security Foundation (OpenSSF)](https://www.openssf.org/) to improve the supply-chain security of the open-source ecosystem.

PR #882, opened 2023-09-14
Set actions.yml to run with read-only permissions

pnacht:
Fixes https://github.com/keras-team/keras-core/issues/881.

This PR ensures all the jobs in the `actions.yml` workflow run with read-only permissions instead of the default `write-all`.


fchollet (approved):
LGTM, thanks!

PR #885, opened 2023-09-14
Don't show optimizer params if not optimizer or not built

mattdangerw:
Most people will call `summary()` before `fit()`. Doing so will usually mean the optimizer would show up as 0 parameters, which is misleading.

### DIFF
--- a/keras_core/utils/summary_utils.py
+++ b/keras_core/utils/summary_utils.py
@@ -311,9 +311,11 @@ def print_summary(
     if model.compiled and model.optimizer and model.optimizer.built:
         optimizer_weight_count = count_params(model.optimizer.variables)
         optimizer_memory_size = weight_memory_size(model.optimizer.variables)
+        optimizer_built = True
     else:
         optimizer_weight_count = 0
         optimizer_memory_size = 0
+        optimizer_built = False
 
     total_count = trainable_count + non_trainable_count + optimizer_weight_count
     total_memory_size = (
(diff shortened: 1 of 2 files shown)


mattdangerw:
Alternately, we could show a `(built)` status.

```
Model: "res_net_2"
┏━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━━┓
┃ Layer (type)                               ┃ Output Shape                        ┃       Param # ┃
┡━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━━┩
│ dense_6 (Dense)                            │ ?                                   │   0 (unbuilt) │
├────────────────────────────────────────────┼─────────────────────────────────────┼───────────────┤
│ dense_7 (Dense)                            │ ?                                   │   0 (unbuilt) │
├────────────────────────────────────────────┼─────────────────────────────────────┼───────────────┤
│ dense_8 (Dense)                            │ ?                                   │   0 (unbuilt) │
└────────────────────────────────────────────┴─────────────────────────────────────┴───────────────┘
 Total params: 0 (0.00 B)
 Trainable params: 0 (0.00 B)
 Non-trainable params: 0 (0.00 B)
 Optimizer params: 0 (0.00 B) (unbuilt)
```

In some ways this is more consistent maybe? But would also mean most people will see an `unbuilt` optimizer in the standard usage and not be sure how to build it.

Seemed potentially least confusing to just leave it off, but open to suggestions!

keras_core/utils/summary_utils.py
```diff
@@ -349,11 +351,13 @@ def print_layer(layer, nested_level=0):
...
-        bold_text(" Optimizer params: ")
-        + highlight_number(f"{optimizer_weight_count:,}")
-        + f" ({readable_memory_size(optimizer_memory_size)})"
-    )
+    if optimizer_built:
+        console.print(
+            bold_text(" Optimizer params: ")
+            + highlight_number(f"{optimizer_weight_count:,}")
+            + f" ({readable_memory_size(optimizer_memory_size)})"
+            + ("" if optimizer_built else " (unbuilt)")
```
qlzh727, line 359: The if condition seems always be True here since the l354 is true.

mattdangerw: oops! good catch, mean to remove this

qlzh727 (approved):

fchollet (approved):
LGTM, thanks!

PR #886, opened 2023-09-14
Export dtype utils in `keras_core.backend`

mattdangerw:
Making multi backend dtype comparisons for code based on keras-core is tricky. Torch has dtype classes such that
`torch_tensor.dtype is string_dtype` will always be false.

We use `standardize_dtype` extensively in our own code and should expose it to clients. `is_float_dtype` and `is_int_dtype` we can include as useful helpers.


fchollet (approved):
LGTM, thanks!

PR #887, opened 2023-09-14
Add missing dtypes.

gleize:
My model contains some `uint` non trainable variables that triggered an exception when calling the `model.summary()` method.

I added the missing dtypes.


fchollet (approved):
LGTM, thanks

PR #912, opened 2023-09-18
[GHA] fix several codecov issues 

kiukchung:
This PR does a few things:

1. Excludes `keras_core/legacy` from coverage metrics since this module there for BC reasons
2. Adds the test backend (jax, tensorflow, torch, numpy) to flags to allow tracking coverage metrics per backend. Note that each backend's coverage will be lower than the aggregate (e.g. `keras_core` will have higher coverage than `keras_core-jax`), so this flag should be used to track trends rather than impose thresholds.
3. Moves `coverage` and `pytest` configuration from `setup.cfg` to `pyproject.toml` to keep all project configs in toml vs cfg (unfortuantely `flake8` configs need to stay in `setup.cfg` since flake8 doesn't support `pyproject.toml` yet: https://github.com/PyCQA/flake8/issues/234)
4. Adds `fail_ci_if_error: true` to codecov upload steps to fail the CI (and force the author to retry the workflow) on transient codecov upload failures of the form 
    ```
     Unable to locate build via Github Actions API. Please upload with the Codecov repository upload token to resolve issue.', code='not_found
    ```
    This happens when we hit codecov API rate limits since we don't use `CODECOV_TOKEN` (see https://github.com/codecov/codecov-action/issues/837). We could circumvent the issue by adding the token as a secret but GH secrets are not visible from forks, implying that contributors will not be able to get coverage diffs for their PRs. To allow this, the only option seems to be to hardcode the token (which seems ok since the token can be setup with limited permissions) but hardcoding tokens is just sketchy...

### DIFF
--- a/.github/workflows/actions.yml
+++ b/.github/workflows/actions.yml
@@ -61,6 +61,7 @@ jobs:
           env_vars: PYTHON,KERAS_BACKEND
           flags: keras_core.applications,keras_core.applications-${{ matrix.backend }}
           files: apps-coverage.xml
+          fail_ci_if_error: true
       - name: Test with pytest
         run: |
           pytest keras_core --ignore keras_core/applications --cov=keras_core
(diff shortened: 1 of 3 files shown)


kiukchung:
will graduate from draft commit after I validate GHA and check codecov reports

fchollet (approved):
LGTM -- thank you!

PR #913, opened 2023-09-18
Refactor merging layers tests as parametrized tests.

hertschuh:
They are almost identical for every merging layer.

This simplifies adding tests that apply to every layer. This is in preparation for adding sparse tensors support.


fchollet (approved):
Nice code savings! LGTM

PR #923, opened 2023-09-19
Adds JAX version checking for Export

nkovela1:
This PR is a temporary fix in response to the incompatibility between the latest JAX release and latest stable TF release, affecting the export API and its use of JAX2TF.


fchollet (approved):
LGTM, thank you!

PR #924, opened 2023-09-19
Fix JAX RNN backend issue.

qlzh727:
This PR address several issues:

1. The existing RNN layer is not training properly due the usage of a fresh StatelessScope in the jax.lax.scan loop. This is causing all the trainable variables to miss the mapping to the actual value in the training loop. Update them to use the parent Stateless scope if it is there. This will address the training issue https://github.com/keras-team/keras-core/issues/322

2. The RNN layers with dropout will have a RNG seed update in the step function, which is not allowed by the jax.lax.scan. We noticed this issue since the updated seed is traced for non-trainable variable, and raise error when we try to put sharding constraint for distribution. Added a new method to pre-populate the dropout mask on the layer and make the inner_loop to be stateless.

3.  During the unit test, I noticed the stackRNNCell doesn't work with existing RNNCell, since it unwrap the list for the state, make the call function to keep the list if the input state is a list.

4. Expose the SimpleRNN|GRU|LSTM cells in the __init__.py since they are public API.

### DIFF
--- a/keras_core/backend/jax/rnn.py
+++ b/keras_core/backend/jax/rnn.py
@@ -1,8 +1,10 @@
+import contextlib
+
 import tree
 from jax import lax
 from jax import numpy as jnp
 
-from keras_core.backend.common.stateless_scope import StatelessScope
+from keras_core.backend.common import stateless_scope
 from keras_core.utils.nest import pack_sequence_as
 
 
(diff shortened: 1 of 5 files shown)


fchollet (approved):
Thanks for the fix!

PR #926, opened 2023-09-20
Fix the inconsistency in `affine_transform` between numpy and jax

james77777778:
This PR resolves the inconsistency in `affine_transform` between numpy and jax due to the unfixed `scipy.ndimage.map_coordinates`.

Previous:
- jax: fixed version of `map_coordinates` in `ops.image.affine_transform`
- numpy: unfixed version of `map_coordinates` in `ops.image.affine_transform`

References:
- https://github.com/google/jax/issues/5687

Let's check if the CI is functioning as expected.

EDITED:
I have found that using `interpolation="nearest"` leads to test failures in torch and jax. The root cause might be differences in the output indices, resulting in significant error in image values.

Currently, I have no idea how to fix it, so `self.skipTest` has been added.

EDITED2:
After the change, I encountered a failure in `nn.moments` in jax. `atol` and `rtol` have been increased to `1e-5` to pass the test.

I believe that the failure now only occurs in `discretization_test.py`.

### DIFF
--- a/keras_core/backend/jax/image.py
+++ b/keras_core/backend/jax/image.py
@@ -58,11 +58,11 @@ AFFINE_TRANSFORM_INTERPOLATIONS = {  # map to order
     "bilinear": 1,
 }
 AFFINE_TRANSFORM_FILL_MODES = {
-    "constant": "grid-constant",
-    "nearest": "nearest",
-    "wrap": "grid-wrap",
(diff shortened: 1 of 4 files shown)


fchollet (approved):
Thanks for the fix!

The test failure is unrelated.

PR #930, opened 2023-09-20
Added dtype to config in Loss.

PatReis:
Thank you very much for the update in `Loss` . 

I have following question/suggestion, which I hope is okay and does not clash with the saving/loading behaviour of models: I added the dtype information to the config of `Loss` such that the default floatx behaviour can be maintained even when created from config. If `dtype` is specified, then the `Loss` will set dtype from config.

If the dtype should not be passed on to the config then this pull request should be canceled.


PatReis:
Ah I see, the tests do not like dtype in config.

PR #932, opened 2023-09-20
Slice serialization

gleize:
Models that contain slicing cannot be serialized because of the missing handling of slices in the `(de)serialize_keras_object` functions.

Thus a simple functional as this one.
```python
x = inp = Input(...)
...
x = x[:, 1:-1, 1:-1]
...
fn = Functional(inp, x)
```
could not be serialized.

Now it's possible with this fix.


fchollet (approved):
Thanks for the fix! LGTM

